PR #82, opened 2026-04-03
feat(rm): add purge support for force delete

overtrue:
## Summary
- add a `rm --purge` flag that sends the RustFS force-delete header for single-object and recursive deletes
- add purge-aware delete helpers in the S3 adapter and cover the request header behavior with unit tests
- add an integration test that verifies normal deletes leave a delete marker while `--purge` permanently removes versioned objects

## Background
RustFS already supports permanent deletion through the `x-rustfs-force-delete: true` request header, but the CLI did not expose that capability. As a result, versioned deletes always produced delete markers even when the backend could perform a permanent delete.

## Validation
- `cargo fmt --all`
- `cargo test --workspace`
- `TEST_S3_ENDPOINT=http://localhost:9000 TEST_S3_ACCESS_KEY=accesskey TEST_S3_SECRET_KEY=secretkey cargo test --package rustfs-cli --test integration --features integration version_operations::test_rm_purge_permanently_deletes_versioned_object -- --exact --nocapture`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,6 +74,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
+name = "assert-json-diff"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


weisd:
https://github.com/rustfs/rustfs/issues/2222
https://github.com/rustfs/rustfs/issues/768

PR #107, opened 2026-04-10
test(cli): cover cors XML alias gaps

overtrue:
## Summary
This PR adds focused regression coverage for two untested CORS paths introduced by the recent bucket CORS feature work.

The first gap was XML normalization when optional header fields are present but blank. The JSON parser path already had coverage for trimming empty `allowedHeaders` and `exposeHeaders`, but the XML path did not. This matters because `rc bucket cors set` accepts both JSON and XML inputs, and users can reasonably provide empty XML elements from exported or hand-edited configs.

The second gap was the deprecated top-level `rc cors get` alias. The newer nested `rc bucket cors get` path already had parser coverage, and `rc cors remove` was covered, but the top-level `get` alias was not. This PR verifies that the compatibility path still resolves to the list command and preserves the bucket argument.

## Validation
I attempted to run `make pre-commit` as requested by the automation, but this repository currently has no `Makefile` and no `pre-commit` target. I therefore ran the documented repository checks directly:

- `cargo fmt --all --check`
- `cargo test --workspace`

All of the above passed.

### DIFF
--- a/crates/cli/src/commands/cors.rs
+++ b/crates/cli/src/commands/cors.rs
@@ -680,6 +680,28 @@ mod tests {
     }
 
     #[test]
+    fn test_parse_cors_configuration_xml_drops_blank_optional_headers() {
+        let config = parse_cors_configuration(
+            r#"
+<CORSConfiguration>
+  <CORSRule>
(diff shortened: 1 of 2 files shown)


overtrue:
Consolidated into #137.

PR #113, opened 2026-04-12
test(rm): cover delete option default paths

overtrue:
## Summary
This PR adds focused regression coverage for the default branch of the recent delete-options work introduced for `rm --purge`.

The recent change added `DeleteRequestOptions` and threaded it through the CLI and S3 delete paths so RustFS-specific purge requests can attach the `x-rustfs-force-delete` header. The force-delete path was already covered, but the default path was not. That left a gap where a future change could accidentally enable the purge header for ordinary deletes without a targeted test catching it.

## Root cause
The original follow-up tests centered on the new behavior when `--purge` is enabled. They did not assert the inverse behavior for the unchanged default flow.

## Fix
This PR adds two narrow tests:

- a CLI-layer unit test that verifies `delete_request_options()` keeps `force_delete` disabled when `--purge` is not set
- an S3 client unit test that verifies batch delete requests omit the RustFS force-delete header when default options are used

These tests stay inside the recently changed delete-options surface and do not alter runtime behavior.

## Validation
I could not run `make pre-commit` because this repository checkout does not contain a `Makefile` or a `pre-commit` target. I ran the documented equivalent checks from `AGENTS.md` instead:

- `cargo fmt --all --check`
- `cargo test --workspace`

### DIFF
--- a/crates/cli/src/commands/rm.rs
+++ b/crates/cli/src/commands/rm.rs
@@ -436,4 +436,21 @@ mod tests {
         let options = delete_request_options(&args);
         assert!(options.force_delete);
     }
+
+    #[test]
+    fn test_delete_request_options_disable_force_delete_without_purge() {
+        let args = RmArgs {
+            paths: vec!["test/bucket/object.txt".to_string()],
+            recursive: false,
+            force: false,
+            dry_run: false,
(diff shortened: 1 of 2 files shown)


overtrue:

PR #121, opened 2026-04-13
test(cli): cover event remove command contracts

overtrue:
## Summary
Recent CLI help/example work covered `event add`, but `event remove` still had an untested contract gap. The noun-first parser path `rc bucket event remove ...` also had no dedicated assertion. That left a regression hole in a recently changed area of the command surface.

This change adds focused coverage for both paths and fixes the missing help example output on the remove subcommand itself. `RemoveArgs` now exposes an `after_help` example block so the top-level and noun-first help text both advertise the remove workflow consistently.

## Root cause
The event command group defined shared help examples at the group level and dedicated examples for `add`, but `remove` had no subcommand-specific `after_help`. Help-contract coverage also skipped both `rc event remove` and `rc bucket event remove`, so the missing examples were never asserted.

## Validation
I first added the new help-contract assertions, which failed because `event remove` help did not include an examples section. After adding the remove-specific `after_help`, the targeted help-contract test passed.

The repository does not contain a `Makefile` or `justfile`, so `make pre-commit` cannot run in this checkout. I ran the equivalent required validations directly instead:

- `cargo fmt --all --check`
- `cargo test --workspace`

### DIFF
--- a/crates/cli/src/commands/event.rs
+++ b/crates/cli/src/commands/event.rs
@@ -22,6 +22,11 @@ Examples:
   rc event add local/my-bucket arn:aws:sns:us-east-1:123456789012:alerts --event delete
   rc bucket event add local/my-bucket arn:aws:lambda:us-east-1:123456789012:function:thumbnail --event put,delete";
 
+const EVENT_REMOVE_AFTER_HELP: &str = "\
+Examples:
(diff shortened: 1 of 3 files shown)


overtrue:
Consolidated into #137.

PR #127, opened 2026-04-15
test(cli): cover noun-first stat and share help

overtrue:
## Summary
This change adds focused contract coverage for the noun-first object commands introduced in the recent bucket/object command-group work. The existing help contract suite already covered `rc object copy` and `rc object show`, but it did not assert discoverability for delegated `rc object stat` and `rc object share` subcommands, and there was no direct parser test for `rc object stat`.

## Root cause
The noun-first command groups were added with thin delegation to the existing command implementations, but only a subset of those delegated paths received fast parser/help coverage. That left a gap where regressions in `rc object stat` or `rc object share` discoverability could slip through even though those commands are part of the user-facing noun-first surface.

## Fix
The help contract test matrix now includes `rc object stat --help` and `rc object share --help`, asserting the usage shape and their key option tokens. The CLI unit tests also now parse `rc object stat local/my-bucket/report.json` and verify it routes to `ObjectCommands::Stat` with the expected path.

## Validation
I first ran focused tests for the touched areas:
- `cargo test -p rustfs-cli --test help_contract nested_subcommand_help_contract`
- `cargo test -p rustfs-cli cli_accepts_object_stat_subcommand`

I then ran the repository's documented checks:
- `cargo fmt --all --check`
- `cargo test --workspace`

I also attempted `make pre-commit` because the automation requested it, but this repository currently has no `Makefile` or `pre-commit` target, so that exact command cannot be executed in this checkout.

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -472,4 +472,20 @@ mod tests {
             other => panic!("expected bucket command, got {:?}", other),
         }
     }
+
+    #[test]
+    fn cli_accepts_object_stat_subcommand() {
+        let cli = Cli::try_parse_from(["rc", "object", "stat", "local/my-bucket/report.json"])
+            .expect("parse object stat");
+
+        match cli.command {
+            Commands::Object(args) => match args.command {
+                object::ObjectCommands::Stat(arg) => {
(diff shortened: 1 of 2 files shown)


overtrue:
Consolidated into #137.

PR #134, opened 2026-04-16
test(cli): cover cors parser alias and force

overtrue:
## Summary

Adds focused parser coverage for recent bucket CORS command surfaces.

The recent CORS command work supports both nested bucket commands and the deprecated top-level `rc cors` form. Existing tests covered several nested command paths, but they did not assert that the top-level `get` alias resolves to the list operation or that the CORS `--force` flag is preserved through nested list parsing.

This PR adds two narrow CLI parser tests in `crates/cli/src/commands/mod.rs`:

- `rc cors get local/my-bucket` maps to `CorsCommands::List`.
- `rc bucket cors list local/my-bucket --force` keeps `force = true`.

No runtime CORS behavior or S3 request code is changed.

## Validation

- `cargo fmt --all --check`
- `cargo test -p rustfs-cli cors`
- `cargo test --workspace`

Note: `make pre-commit` was requested, but this checkout has no Makefile and `make` reports `No rule to make target pre-commit`.

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -440,6 +440,19 @@ mod tests {
     }
 
     #[test]
+    fn cli_accepts_top_level_cors_get_alias() {
+        let cli = Cli::try_parse_from(["rc", "cors", "get", "local/my-bucket"])
+            .expect("parse top-level cors get");
+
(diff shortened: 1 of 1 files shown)


overtrue:
Consolidated into #137.

PR #271, opened 2026-07-21
feat(transfer): add multi-source planning and global controls

cxymds:
Closes rustfs/backlog#1372

Parent roadmap: rustfs/backlog#1361

## Background and user impact

`rc cp` accepted only one source and recursive transfers used command-local sequencing. Filtering, retries, concurrency, failure aggregation, and summaries had no shared contract for bulk copy or future mirror workflows.

## Solution

- Adds reusable storage-independent `TransferPlan`, `TransferSelection`, and `TransferExecutor` primitives in `rc-core`.
- Accepts multiple sources and rejects ambiguous or colliding destinations before execution.
- Adds include/exclude globs, strict UTC newer/older boundaries, rewind selection, and configurable empty-selection behavior.
- Enforces command-wide concurrency, aggregate rate pacing, transient retry budgets, fail-fast cancellation, and continue-on-error behavior.
- Preserves deterministic item outcomes and aggregate counters.
- Keeps the legacy single-object path and output contract unchanged unless planner controls are explicitly requested.

## Review resolution

- Loads aliases once per command and creates one reusable `Arc<S3Client>` per distinct alias for planning and execution, preserving TCP/TLS connection pools across objects.
- Removes per-item `AliasManager`, config parsing, AWS config resolution, and client construction from concurrent workers.
- Keeps missing aliases as deterministic item-level failures so `--continue-on-error` summaries retain their prior semantics.
- Uses `tokio::fs::metadata` in planned uploads.
- Reuses `TransferCandidate::size_bytes` for remote-copy size guards and result fallback, eliminating the duplicate HEAD request.
- Represents concurrency and retry flags as `Option` values, so explicitly passing documented defaults still selects the planner while omitted flags preserve the legacy path.
- Adds tests for explicit-default routing, alias deduplication, shared client-pool identity, and candidate-size multipart guards.
- Retains O(N) candidate materialization for this PR because deterministic sorting, collision preflight, and full-plan summaries currently depend on it; streaming planner work should be isolated before mirror parity.

## Scope and deferred work

- Recursive remote-to-remote and multipart copy remain in rustfs/backlog#1371 scope.
- Mirror parity consumes the shared planner in the dependent mirror PR.
- Planned bulk JSON remains explicitly rejected until its versioned batch contract is defined.

## BREAKING

This PR updates the protected `docs/reference/rc/cp.md` behavior contract. The CLI change is additive and backward-compatible: existing one-source syntax and legacy single-object JSON remain valid. No config migration or change to output v1/v2 is required.

## Validation

- `cargo clippy --workspace -- -D warnings`: passed with zero warnings.
- `cargo test --workspace`: passed after integration with latest `main`.
- Compatibility script unit tests and repository consistency check: passed.
- Focused cp tests: 37 passed; transfer-planner integration: 5 passed.
- Launcher, watch, ops, replication, schema-v3, help-contract, and golden coverage retained after semantic merge.


overtrue (changes requested):
## Review: Request changes

Solid, well-abstracted planning/concurrency layer with unusually strong tests (bounded parallelism, cancellation, partial-failure, deterministic exit code all covered), clean layering, and no prohibited patterns. `RateGate` correctly drops its mutex before `sleep_until`. The one substantive problem is that the execution path defeats connection reuse and does blocking I/O on the shared executor task — undercutting the very throughput this feature exists to provide.

### Findings

**[MAJOR] `crates/cli/src/commands/cp.rs:550` (`create_leaf_s3_client`, called per candidate at :491/:515/:531) — a fresh `S3Client` and `AliasManager` are built for every object**
- **Why:** For a recursive/multi-source copy of N objects this constructs N independent connection pools (no TCP/TLS keep-alive across objects) and N full `aws_config::defaults().load().await` resolutions, plus a blocking `std::fs::read_to_string` + TOML parse (`AliasManager::get()` → `ConfigManager::load()`) per object. Because all in-flight transfers share a single task (`FuturesUnordered`, not spawned), each object's blocking config read stalls every other concurrent transfer. This is a regression versus the legacy `download_prefix`/`upload_directory` paths, which build one client and reuse it, and it negates the concurrency/rate benefit this PR adds.
- **Fix:** Construct clients once per distinct alias before `execute()` (or memoize into an `Arc<Mutex<HashMap<String, Arc<S3Client>>>>`) and hand a shared `Arc<S3Client>` to each candidate; drop the per-op `AliasManager::new()`.

**[MINOR] `crates/cli/src/commands/cp.rs:481` — `perform_planned_upload` calls synchronous `std::fs::metadata(source)?` inside the concurrent executor task**
- **Why:** A blocking syscall on the shared executor task freezes all other in-flight transfers for its duration; inconsistent with `perform_planned_download`, which correctly uses `tokio::fs`.
- **Fix:** Use `tokio::fs::metadata`.

**[MINOR] `crates/cli/src/commands/cp.rs:958` (`build_remote_candidates`) — the whole object set is materialized in memory before any transfer starts**
- **Why:** The listing loop accumulates every page into `candidates`, so memory is O(N) for the prefix; the legacy `download_prefix` streamed page-by-page (≤1000 held at a time). For the large-bucket/mirror use cases this planner targets, this is a scalability cliff.
- **Fix:** Acceptable for the current scope, but the planner can't stream — consider a chunked/streaming plan before mirror parity.

**[MINOR] `crates/cli/src/commands/cp.rs:532` — `perform_planned_remote_copy` re-issues `head_object`**
- **Why:** `build_remote_candidates` already did a HEAD (:1003) and stored `size_bytes` on the candidate, so this is an extra round-trip per S3-to-S3 object.
- **Fix:** Reuse `item.size_bytes` for the 5 GiB guard.

**[NIT] `crates/cli/src/commands/cp.rs:275` (`uses_transfer_planner`) — planner routing compares flags against hard-coded defaults (`concurrency != 4`, `retry_attempts != 3`, …)**
- **Why:** Passing an explicit default (e.g. `--concurrency 4`) silently takes the legacy path and ignores the control; also desyncs if a default constant changes in one place only.
- **Fix:** Track "was set" via `Option`/clap value-source rather than value comparison.

### Test coverage

Strong for the executor (global concurrency bound, aggregate rate limiting, transient-only retry, fail-fast cancellation, drain-but-don't-schedule partial failure) and planner integration. Gaps: no test asserts client/connection reuse across objects (would catch the MAJOR); `--older-than`/`--rewind` are covered at unit level only.

### Nice work

`TransferExecutor` fail-fast semantics are well-reasoned: cancellation only stops *scheduling*, in-flight work settles, and `first_failure()` returns the lowest-plan-index failure so the exit code is deterministic and cannot be masked by a cancelled higher-index item. The protected `docs/reference/rc/cp.md` change is properly documented with a BREAKING section.

cxymds:
All requested changes are implemented on the latest main: one shared S3 client/config per alias, async metadata, no duplicate remote HEAD, and explicit-default flags now route through the planner. Regression tests cover client-pool reuse, routing, and size guards; the full mandatory workspace suite passes. O(N) materialization remains the explicitly accepted deferred item. Please re-review.

PR #272, opened 2026-07-21
feat(cli)!: add runtime discovery and replication workflows

cxymds:
## Related issues

Closes rustfs/backlog#1362.
Closes rustfs/backlog#1363.
Closes rustfs/backlog#1385.
Closes rustfs/backlog#1409.
Closes rustfs/backlog#1410.
Closes rustfs/backlog#1418.
Closes rustfs/backlog#1419.

Roadmap context: rustfs/backlog#1361, rustfs/backlog#1381, and rustfs/backlog#1411.

## Background

This PR is the final integration point for the reviewed CLI roadmap stack. It combines the output-v3 foundation, runtime feature discovery, bounded replication operations, safe site-replication workflows, and the explicit-stack command launcher required by large AWS SDK futures.

## Summary

- add the versioned output-v3 contract, fixtures, compatibility tests, and migration guidance while retaining output v1 and v2
- add typed RustFS runtime capability discovery with process-scoped, credential- and TLS-context-aware caching
- add bounded replication diff, active target checks, and bucket resync start/status workflows
- add bounded site-replication edit and persisted resync start/status/cancel workflows with explicit confirmation and unknown-outcome handling
- preserve opaque server fields where required while sanitizing terminal output and preventing credential disclosure
- run command futures and Tokio workers on documented 16 MiB stacks
- preserve default panic diagnostics while mapping command-thread failures to a stable exit
- shut down the runtime in the background so wedged blocking tasks cannot stall CLI exit
- include unit, integration, process, wire-contract, golden-fixture, schema, and launcher coverage

## BREAKING change and migration

This integration extends the protected CLI reference contract and introduces output v3. Existing output-v1 and output-v2 schemas remain unchanged. New replication commands use output-v3 families, and site-replication info migrates from raw server JSON to the documented v3 `admin_operations` envelope. Migration guidance is included in `docs/reference/rc/`.

## Safety and compatibility

- request, success, and error bodies are bounded for the new workflows
- redirects and automatic mutation retries are disabled where replay could duplicate effects
- active mutations require confirmation before alias or network setup
- persisted site-resync snapshots are not presented as live lifecycle state
- credentials, private keys, and unknown sensitive server fields are not emitted
- core and CLI remain independent of AWS SDK crates

## Validation

- `cargo fmt --all --check`
- `cargo test --workspace`

All checks passed locally on the final tree after integration with the latest `origin/main`. GitHub Actions validates the same head before merge.


overtrue (changes requested):
## Review: Request changes

The core idea — move the async command graph onto a named worker thread with a 16 MiB stack to avoid the Windows main-thread stack limit — is sound, and the launcher is cleanly written and well-tested. The generics are also correct: the future is only `'static` (not `Send`) and is constructed on the worker thread, so the large AWS SDK future never crosses the thread boundary. My concern is the accompanying panic hook, which trades away all crash diagnostics with no escape hatch, plus two coverage gaps in what the fix actually protects.

### Findings

**[MAJOR] `crates/cli/src/main.rs:65` (`install_command_panic_hook`) — command-thread panics are fully silenced (message, location, and backtrace) with no escape hatch**
- **Why:** On any ordinary panic in the command future the hook returns without calling `default_hook`, so even `RUST_BACKTRACE=1` prints nothing; the user/CI gets only `Command execution thread terminated unexpectedly` + exit 1, and nothing is logged via `tracing`. Note the original #1385 trigger — an actual *stack overflow* — aborts the process (SIGABRT) and never reaches `join()`/`ThreadPanicked`, so this hook only discards diagnostics for ordinary bugs while the 16 MiB stack bump alone fixes the real issue. Net effect: a debuggability regression versus `#[tokio::main]`.
- **Fix:** Honor `RUST_BACKTRACE` (or an `RC_*` var), or call `default_hook` unconditionally and keep mapping the `join()` `Err` → `ThreadPanicked` exit (the exit-code mapping is independent of whether the hook prints). Optionally `tracing::error!` the payload before exiting.

**[MINOR] `crates/cli/src/main.rs:65` vs `crates/cli/src/commands/mirror.rs:366,418,500` — panic suppression is inconsistent; only the `rc-command` thread is silenced**
- **Why:** Parallel mirror copy/remove tasks run on `tokio-runtime-worker` threads, so their panics still reach `default_hook` (full payload + backtrace) and are additionally formatted into `"Mirror … worker failed: {join_error}"`. The "without propagating panic payloads" guarantee does not hold for the large-transfer path it targets.
- **Fix:** Enforce clean output uniformly at the task boundary rather than by thread name (or drop the suppression per the MAJOR fix).

**[MINOR] `crates/cli/src/main.rs:79` (`new_multi_thread`, no `.thread_stack_size(...)`) — the 16 MiB stack covers only the top-level `block_on` future, not spawned tasks**
- **Why:** The runtime builder never sets `thread_stack_size`, so mirror's copy/remove multipart futures run on Tokio's default (~2 MiB) worker stacks regardless of this fix. If an overflow lives in a spawned multipart future, mirror stays unfixed. The `command_launcher_constructs_and_polls_large_future_on_worker` test uses a `current_thread` runtime, so it exercises only the `block_on` path and cannot catch this.
- **Fix:** Also set `.thread_stack_size(COMMAND_THREAD_STACK_SIZE)` on the runtime builder, or confirm heavy work never runs on spawned tasks.

**[MINOR] `crates/cli/src/main.rs:98,110` — the runtime is now gracefully dropped (blocking shutdown-join) before `process::exit`, a behavior change from `#[tokio::main]`**
- **Why:** The old path called `process::exit` *inside* `block_on`, bypassing runtime drop entirely (instant exit). Every command now pays a shutdown-join, and a wedged blocking task could stall exit where it previously returned immediately.
- **Fix:** If instant exit is intended, use `shutdown_background()` / `shutdown_timeout(..)` to bound it.

### Test coverage

Six tests cover the launcher mechanism well (exit-code preservation, named-thread placement, a 2 MiB future, runtime-failure mapping, spawn-failure message, panic→`ThreadPanicked`). Gaps: `install_command_panic_hook` behavior itself is untested, and the production `multi_thread` + worker-stack path is never exercised (the large-future test uses `current_thread`).

### Note

The panic-hook silencing interacts with #273 (watch): a `BrokenPipe` panic on streaming output unwinds this thread and surfaces as `ThreadPanicked` + a nonzero exit with no explanation. Worth fixing together.

PR #273, opened 2026-07-21
feat(watch): stream RustFS object notifications

cxymds:
## Related issues

- Closes rustfs/backlog#1374
- Depends on #268
- Depends on #266
- Depends on #272

## Background and user impact

`rc` had no equivalent to `mc watch`, and the existing request paths were designed around bounded responses rather than a notification body that remains open indefinitely. Users therefore could not observe RustFS object creation, access, removal, version, or delete-marker activity from the CLI without polling.

## Root cause

The core/S3 boundary did not expose a typed streaming notification API, the S3 transport had no incremental decoder for RustFS ListenNotification frames, and the CLI/output layers had no cancellation, reconnect, or JSON Lines path for a long-lived command.

## Solution

- Add a typed `WatchApi`, request, frame, event, source metadata, and boxed stream contract in `rc-core`.
- Add a SigV4-signed RustFS beta.10 ListenNotification transport in `rc-s3` for root and bucket routes, repeated event filters, prefix/suffix/ping parameters, bounded incremental JSON-line decoding, transport keepalives, URL-decoded keys, version/delete-marker metadata, and stable HTTP error categories.
- Add `rc watch ALIAS[/BUCKET]` with mc-compatible `put`, `delete`, and `get` shorthands, compact schema-v3 JSON Lines, terminal-safe human output, quiet mode, bounded exponential reconnects, and prompt Ctrl-C exit code 130.
- Advertise `listen_notification` only for the exact known RustFS `1.0.0-beta.10` release family while leaving undeclared future-server support authoritative to the signed route response.
- Extend schema v3 additively with optional watch source and delete-marker fields and document the new command.

## BREAKING protected-file process

**BREAKING:** this PR adds a new public command contract under the protected `docs/reference/rc/` tree. The behavior is additive: no existing command, flag, exit code, configuration schema, or v1/v2 output contract changes. `schemas/output_v1.json` and `schemas/output_v2.json` are byte-for-byte unchanged relative to the stacked base.

Existing consumers require no migration. New watch consumers should read one schema-v3 JSON object per line, dispatch on `schema_version`/`type`, and tolerate optional `delete_marker` and `source` fields. Transport keepalives are intentionally suppressed.

## Validation

- `cargo fmt --all --check`
- `CARGO_BUILD_JOBS=1 CARGO_INCREMENTAL=0 cargo clippy --workspace --all-targets --target-dir /tmp/rc-target-1374 -- -D warnings`
- `CARGO_BUILD_JOBS=1 CARGO_INCREMENTAL=0 cargo test --workspace --target-dir /tmp/rc-target-1374`
  - rc-core: 128 passed
  - rc-s3: 180 passed
  - CLI: 400 passed
  - all integration, schema, help-contract, and doc tests passed
- `CARGO_BUILD_JOBS=1 CARGO_INCREMENTAL=0 INSTA_UPDATE=no cargo test -p rustfs-cli --features golden --test golden --target-dir /tmp/rc-target-1374` (5 passed)
- Focused watch tests: rc-core 2 passed, rc-s3 7 passed, CLI 28 passed across its lib/main harnesses
- `git diff --check`
- v1/v2 schema diff against `origin/cxymds/issue-1363-capabilities`: no changes
- v3 schema and JSONL fixtures parse successfully with `jq`
- dependency-boundary, `unsafe`, production `.unwrap()`, and credential-output scans: clean

Head commit: `5033905054aabd6cb8071a68dcc3de8c3df35be8`


overtrue (changes requested):
## Review: Request changes

The streaming core is genuinely strong: an incremental chunk-split decoder with bounded per-line/per-error buffers, faithful reproduction of the MinIO/RustFS transport (text/event-stream content-type but NDJSON body with whitespace/empty-`Records` keepalives), terminal-injection sanitization, capability gating that refuses to infer support for undeclared servers, and thorough unit tests. One real defect blocks it — the most idiomatic way to consume the stream panics — plus two long-lived-stream behaviors worth reconsidering.

### Findings

**[MAJOR] `crates/cli/src/output/formatter.rs:480` (`json_line`) & `:506` (`println`), driven by `watch.rs:464/476` — streaming output uses `println!`, which panics on `BrokenPipe`**
- **Why:** Rust ignores SIGPIPE by default and rc never resets it. `rc watch local/ --json | head -n5` (or `| grep -m1`, or quitting `| less`) closes stdout while the infinite stream keeps writing; the next `println!` panics (`failed printing to stdout: Broken pipe`). The panic unwinds the `rc-command` thread (the `main.rs` panic hook only *silences* the message, it does not stop unwinding) → `join()` `Err` → `LauncherError::ThreadPanicked` → `Command execution thread terminated unexpectedly` on stderr and a nonzero exit. So the most natural way to consume a stream yields a spurious error and the wrong exit code. Affects both `--json` and human paths.
- **Fix:** Treat `ErrorKind::BrokenPipe` as clean termination in the output layer (exit 0 or 130 instead of panicking), or reset SIGPIPE to `SIG_DFL` at startup. A scoped guard in `json_line`/`println` is sufficient.

**[MINOR] `crates/cli/src/commands/watch.rs:276` (`run_watch_loop`) — `reconnects` is a lifetime counter, never reset after a healthy connection**
- **Why:** It is initialized to 0, only ever incremented (:291, :319), and the backoff is derived from it. A watch that streams fine for hours but hits its Nth (default 5) *unrelated* transient drop over its whole lifetime exits permanently, even though each drop individually recovered; by the last incident the backoff is already pinned at `--reconnect-max-delay-ms`. This departs from `mc watch`'s indefinite retry and is surprising for a monitoring tool.
- **Fix:** Reset `reconnects = 0` after a frame is successfully received, so `--reconnect-attempts` counts *consecutive* failures.

**[MINOR] `crates/s3/src/watch.rs:258` (`decode_line`) + `watch.rs:316` — a single malformed event line is fatal (no reconnect), ending the whole watch**
- **Why:** For a long-lived monitor, one bad frame (partial write, proxy injection, server bug) permanently kills monitoring. This is intentional and documented/tested — flagging so the tradeoff is explicit.
- **Fix:** Consider emitting an error record and continuing instead of aborting.

### Test coverage

Strong for the decoder (chunk-splitting, keepalive classification, URL-decoded keys, delete-marker, malformed surfacing), SigV4 signing, status→exit mapping, backoff math, disconnect+reconnect+cancellation, exhausted-retry, schema validation, and sanitization. Gaps: no broken-pipe/EPIPE test, no reset-after-healthy scenario (because it isn't reset), and no full `execute()` integration against a mock streaming HTTP server.

### Nice work

`docs/reference/rc/output.md` was updated so schema/docs/code stay consistent on the keepalive contract (rc consumes transport keepalives without emitting a line; event records always set `keepalive:false`).

PR #275, opened 2026-07-21
feat(rc-ops)!: add health, readiness, and usage commands

cxymds:
Refs rustfs/backlog#1377

## Background and user impact

The CLI did not expose RustFS liveness/readiness probes or the server's background-scanner data-usage snapshot. Operators had to combine curl, admin credentials, and custom S3 listing scripts, and a client-side usage scan could become unexpectedly expensive if started implicitly.

The root cause was the absence of typed health and usage boundaries in `rc-core`, RustFS adapters in `rc-s3`, and schema-v3 command families for these operational results.

## What changed

- Adds `rc ping` for bounded unsigned `GET /health` liveness checks with latency and endpoint identity.
- Adds `rc ready` for bounded unsigned `GET /health/ready` dependency-readiness checks.
- Adds `rc du ALIAS[/BUCKET[/PREFIX]]`.
- Uses the signed RustFS `GET /rustfs/admin/v3/datausageinfo` snapshot when capability discovery advertises `admin.data-usage`.
- Requires explicit `--fallback` before any slower client-side S3 scan.
- Supports paginated current objects, versions/delete markers, incomplete multipart uploads, and uploaded parts according to `--versions` and `--incomplete`.
- Reports source, snapshot timestamp/age/staleness, totals, per-bucket rows, partial results, and sanitized failures in human and JSON-v3 output.
- Preserves authentication denial unless the user explicitly permits fallback.

## RustFS compatibility evidence

The implementation was checked against RustFS `main` and the beta.10 contract:

- `GET /health` and `GET /health/ready` are public operational endpoints.
- `GET /rustfs/admin/v3/datausageinfo` is registered as the signed data-usage route.
- RustFS `DataUsageInfo` supplies last-update, object/version/delete-marker totals, bytes, and per-bucket usage.
- Prefix and incomplete-upload dimensions are not present in the snapshot, so those requests require explicit S3 fallback.
- Capability hardcoding is exact for `1.0.0-beta.10`; later versions may use advertised discovery instead of an over-broad version match.

## Safety and failure semantics

- Health timeouts accept only 1–300 seconds and cover the request/response operation.
- HTTP 429/503 readiness is unhealthy and exits with retryable network code 3.
- Permission failures map to authentication code 4 and never start a scan without `--fallback`.
- Alias-wide scan failures preserve successful buckets, mark output partial, and exit 3.
- Pagination guards reject truncated pages that do not advance their markers.
- Totals use saturating arithmetic; server-controlled text is sanitized before human output.

## BREAKING contract note

This PR adds new command contracts under the protected `docs/reference/rc/` path and is therefore marked BREAKING per repository policy. Existing commands, exit codes, config schema, and output v1/v2 remain unchanged. The additions use output schema v3; consumers should select the v3 health/usage family for these new commands.

## Dependency chain

1. #272 — command-stack launcher (base: `main`).
2. #266 — output schema v3 (base: #272).
3. #268 — capability discovery (base: #266).
4. This PR — `cxymds/issue-1377-ops-utilities` at `ef327d3d` (base: #268).

## Validation

- `cargo clippy --workspace --all-targets -- -D warnings`: passed with zero warnings.
- `cargo test --workspace -- --format terse`: passed (1,173 tests, 0 failed, 1 existing ignored doctest).
- Core operations contract: 3/3 passed.
- S3 operations adapter tests: 9 matched tests passed.
- CLI focused tests: ping 4/4, ready 3/3, du 6/6 in both library and binary targets.
- Output schema v3: 5/5 passed.
- Help contract: 3/3 passed.


overtrue (changes requested):
## Review: Request changes

High-quality ops commands (`ping`/`ready`/`du`) with clean `?`-based error handling, saturating arithmetic, negative-byte clamping, terminal-escape sanitization of server-controlled strings, unsigned health probes (no credential leakage), and solid exit-code + schema tests. One robustness defect on the most-exercised scan path holds it back, plus a couple of minor gaps.

### Findings

**[MAJOR] `crates/s3/src/ops.rs:289` (`scan_current_objects`) — unguarded pagination can infinite-loop**
- **Why:** The default `rc du --fallback` object-count loop advances `continuation_token = page.continuation_token` with no guard: if a truncated page returns `None` or an unchanged token, it re-lists page 1 forever (client hang / unbounded requests). This directly contradicts the PR description's stated invariant ("Pagination guards reject truncated pages that do not advance their markers") and is inconsistent with the two sibling scanners that *do* guard — `scan_versions` (:342) and `scan_incomplete_uploads` (:400). RustFS is a custom MinIO-derived server, and this is the most-exercised scan path.
- **Fix:** After `if !page.truncated { return }`, require a non-empty `page.continuation_token` that differs from the previous one, else return `Error::Network(...)` exactly like the version/multipart scanners.

**[MINOR] Workspace version unchanged at `0.1.24` despite a protected-path change (`docs/reference/rc/ops.md` added, `README.md`/`output.md` edited) + BREAKING marker**
- **Why:** The protected-file policy for `docs/reference/rc/` calls for a version bump alongside the BREAKING marker; the changes are additive, so functional impact is low, but the CI breaking-change gate may reject it.
- **Fix:** Confirm the bump is carried at the stack base/integration, or add it here.

**[MINOR] `crates/cli/src/commands/du.rs` tests — no exit-code test for the `unsupported_feature` path (exit 7)**
- **Why:** Requesting a prefix/`--incomplete` without `--fallback`, or capability unavailable, is the canonical "unsupported" scenario; only the envelope shape is unit-tested, not `du`'s exit code end-to-end. (Success/AuthError/NetworkError are covered, so the ≥2 rule is met.)
- **Fix:** Add one `execute_with_apis` case asserting `ExitCode::UnsupportedFeature` and that the scan adapter is not called.

**[NIT] `crates/cli/src/commands/ping.rs:1154` — readiness "not-ready" and transport "error" both map to exit 3 / `network_error`, differing only in message**
- **Why:** This is a correct design for k8s/scripting (any nonzero = not-ready, retryable), but a consumer cannot distinguish "server said not-ready" from "server unreachable" by exit code or type. Flagging only so the intent is explicit.

### Test coverage

Adequate for the exit-code contract and schema (ping 0/3, ready 0/3, du 0/4/3; pagination for objects/versions/multipart; snapshot parsing; auth-vs-unsupported routing). Gaps: du's unsupported→7 path and the snapshot bucket-absent NotFound→5 path are untested; no test exercises a truncated-without-token page (which would have caught the MAJOR).

### Nice work

No client scan is ever started after an unauthorized/unsupported admin request unless `--fallback` is explicit (auth denial preserves exit 4), and all server-controlled identity/bucket/failure strings are escaped in human output to prevent terminal-escape injection.

PR #287, opened 2026-07-21
feat(rc-replication)!: add a bounded prefix replication diff

cxymds:
Closes rustfs/backlog#1409

## Background

RustFS beta.10 exposes an on-demand replication diff route, but rc had no typed or bounded client workflow for it. Operators could not inspect pending or failed object versions without reaching for raw Admin API requests.

## Solution

- add `rc replicate diff` and `rc bucket replication diff` with an optional prefix
- call the exact signed `POST /rustfs/admin/v3/replication/diff` route with an empty body
- bound both success and error responses to 8 MiB
- preserve future server fields while strictly validating required scan fields
- distinguish unsupported routes from missing buckets or replication configuration
- add deterministic human/JSON output, output-v3 schema fixtures, process tests, and protected reference documentation
- report truncated scans as partial and explicitly non-resumable

This PR intentionally does not claim time filtering, resume tokens, replication metrics, MRF truth, or repair behavior.

## BREAKING CHANGE

The protected CLI reference contract is extended with the new replication diff command and the output-v3 contract gains the `replication` family. Existing output-v1 and output-v2 contracts are unchanged.

## Validation

- `cargo fmt --all --check`
- `cargo test --workspace`
- independent source/protocol review against RustFS beta.10 and current main

All local checks passed. Live container validation was not run because no Docker daemon is available in the current environment.


cxymds:
Validation update for `e4243986cc884f578dcd94f15c0249c92bc54645`:

- CI: https://github.com/rustfs/cli/actions/runs/29860937581 — passed on Linux, macOS, and Windows
- Integration Tests: https://github.com/rustfs/cli/actions/runs/29860938139 — passed
- local formal gates: fmt, workspace/all-target Clippy with zero warnings, and workspace tests passed
- independent protocol/code review: no actionable findings

The first Windows CI attempt exposed a test-only socket hang in the chunked overflow fixture. The fixture now sends exactly 8 MiB + 1 byte, bounds socket writes and completion waits, and retains the production response-limit assertion. The repaired test passed 10 serial repetitions locally, the full rc-s3 suite, full workspace gates, and the replacement Windows CI job. Production response handling was not changed.

Live RustFS container validation remains unavailable because the current environment has no Docker daemon.

PR #351, opened 2026-08-24
fix(mirror): skip recopies when source ETag is preserved in metadata

### LINKED ISSUE #342: `rc mirror` S3-to-S3 does not preserve ETag, causing unchanged objects to be copied again
## Description

I am using `rc 0.1.31` to mirror objects between two RustFS S3 endpoints:

```text
RustFS source (remote NTFS) → RustFS destination (local NTFS)
```

The problem appears to be related to how `rc mirror` performs S3-to-S3 copies rather than to the underlying filesystem.

### Observed behavior

I initially ran:

```bash
time rc mirror --overwrite --remove --json \
  remoto/archive/2026/07 \
  local/archive/2026/07
```

Result:

```json
{
  "source": "remoto/archive/2026/07",
  "target": "local/archive/2026/07",
  "copied": 764,
  "removed": 0,
  "skipped": 0,
  "errors": 0,
  "dry_run": false
}
```

The operation took approximately 30 minutes.

Running the exact same command again, without any changes on the source, still caused all 764 objects to be copied again.

A dry run confirmed:

```json
{
  "copied": 764,
  "removed": 0,
  "skipped": 0
}
```

## ETag difference

For example, one object had:

### Source

```text
Size: 21375369 bytes
ETag: 513cf08fa7a370c31e6cee7beb120163-2
Date: 2026-08-04 22:11:52 UTC
```

### Destination after `rc mirror`

```text
Size: 21375369 bytes
ETag: 53d0b16993c80d285b6e681ef17ef6b3-1
Date: 2026-08-11 13:28:51 UTC
```

The content size is identical, but the ETag is different.

This causes the next `rc mirror` invocation to consider the object changed and copy it again.

## Important comparison with `mc --preserve`

I then used `mc cp --preserve` on a smaller subset of the same S3-to-S3 data.

After that, the same object reported:

### Source

```text
Size: 21375369 bytes
ETag: 513cf08fa7a370c31e6cee7beb120163-2
Date: 2026-08-04 22:11:52 UTC
```

### Destination

```text
Size: 21375369 bytes
ETag: 513cf08fa7a370c31e6cee7beb120163-2
Date: 2026-08-11 14:17:36 UTC
```

Notice that the `Date`/mtime is still different, but the ETag is identical.

I then ran:

```bash
rc mirror --overwrite --remove --json \
  remoto/archive/2026/07/20260726/D7500 \
  local/archive/2026/07/20260726/D7500
```

and got:

```json
{
  "copied": 0,
  "removed": 0,
  "skipped": 14,
  "errors": 0,
  "dry_run": false
}
```

The operation took only:

```text
real    0m0.330s
```

This strongly indicates that `rc mirror` uses the ETag (rather than the object modification date) as an important part of its equality check.

## Expected behavior

For an S3-to-S3 mirror, if the source object is copied without changing its content, I would expect the destination object to retain an identity that allows subsequent mirror operations to recognize it as unchanged.

Ideally:

```text
source ETag == destination ETag
```

after a successful mirror operation, when the destination object contains exactly the same object data.

Otherwise, every subsequent `rc mirror` operation unnecessarily transfers the entire object again.

## Why this matters

My use case is a RustFS server acting as a continuously updated mirror of another RustFS server.

The current behavior turns an incremental mirror into a full re-transfer.

After copying the same objects with `mc --preserve`, subsequent `rc mirror` correctly skips them.

Therefore, preserving object identity/ETag during S3-to-S3 mirroring would make a very significant difference for this use case.

## Environment

```text
rc version: 0.1.31
Source: RustFS S3 (rustfs 1.0.0-rc.1)
Source filesystem: NTFS3
Destination: RustFS S3 (rustfs 1.0.0-rc.1)
Destination filesystem: NTFS3
Transfer type: S3-to-S3
```

loverustfs:
## Related issue

Fixes https://github.com/rustfs/cli/issues/342

## Background

`rc mirror` copies S3-to-S3 objects through a download-then-upload path. Multipart completion on the destination stores a different ETag than the source (`hash-2` vs `hash-1` in the report). The planner compared ListObjects ETags, so a second `--overwrite` run recopied every object even when the bytes were unchanged.

`mc cp --preserve` keeps the source ETag via server-side copy. Cross-alias mirror cannot do that, and ListObjects does not return user metadata, so ETag-only planning cannot see a client-recorded identity.

## Solution

- Remote uploads that have a source ETag now write `x-amz-meta-rc-source-etag`.
- `--compare auto|etag|size` (default `auto`):
  - `auto`: skip when ETags match, or when size matches and destination metadata records the source ETag
  - `etag`: skip only on identical stored ETags
  - `size`: skip when sizes match
- In `auto` mode, destinations with the same size but a different listed ETag are inspected with HeadObject so the identity metadata can be read.
- Destination race/removal checks ignore identity metadata so List vs HEAD differences do not look like a concurrent change.

This is additive. Existing flags and default skip-on-matching-ETag behavior are unchanged. Objects copied before this change still recopy once; after that they skip.

This PR must be marked BREAKING because `docs/reference/rc/mirror.md` is a protected CLI behavior contract. The change is additive; no JSON schema or config `schema_version` bump applies.

## Tests

- Unit: Auto/ETag/Size skip rules, missing or mismatched identity recopies, size mismatch never skips, case-insensitive identity metadata, HeadObject lookup gating, identity ignored in race checks, upload records the source ETag
- Integration: dry-run remote-to-remote Auto skip after HEAD, ETag mode recopies without HEAD, Auto recopies when identity is missing or wrong, size mismatch recopies without HEAD
- Help contract includes `--compare`
- Regression: `scripts/regression/mirror-identity.sh`

## Validation

```bash
cargo fmt --all
cargo test --workspace
./scripts/regression/mirror-identity.sh
```

### DIFF
--- a/crates/cli/src/commands/mirror.rs
+++ b/crates/cli/src/commands/mirror.rs
@@ -1,6 +1,7 @@
 //! mirror command - Synchronize trees between local filesystems and S3-compatible storage.
 
 use std::collections::{BTreeMap, HashMap};
+use std::future::Future;
 use std::path::{Component, Path, PathBuf};
 use std::sync::Arc;
 
@@ -495,7 +496,12 @@ impl LiveMirrorIo {
             ));
         };
 
-        match self.target_disposition(operation).await? {
+        let disposition =
+            source_preflight_then_target(source_client.as_ref(), source_path, operation, || {
+                self.target_disposition(operation)
+            })
+            .await?;
+        match disposition {
             TargetDisposition::AlreadyComplete => {
                 return Ok(operation.source.snapshot.size_bytes.unwrap_or_default());
             }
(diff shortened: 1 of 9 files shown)


crates/cli/src/commands/mirror.rs
```diff
@@ -542,14 +604,16 @@ impl LiveMirrorIo {
     ) -> rc_core::Result<TargetDisposition> {
         let current = self.target.current_entry(&operation.relative_path).await?;
         if let Some(current) = &current
-            && source_matches_target(&operation.source, current)
+            && source_matches_target(&operation.source, current, self.compare)
```
cxymds: [P1] The identity shortcut runs before the planned destination snapshot is checked. If a target is replaced after planning but keeps (or is given) rc-source-etag equal to the planned source ETag, source_matches_target returns true and this method reports AlreadyComplete without checking expected.snapshot.content_matches. For remote-to-remote copies this also happens before the source HEAD revalidation, so a source that changed after planning can be silently skipped. Please make the identity shortcut conditional on the planned/live snapshot still matching, or use an equivalent conditional write, and add a test for this call order.

cxymds: Fixed in 8e59a2a. Remote-to-remote mirror execution now revalidates the planned source snapshot before reading target disposition. The identity shortcut additionally requires the live destination to match the planned content snapshot; otherwise it returns a conflict. Added tests proving source preflight runs before target identity checks and covering replaced or appeared destinations. All workspace checks pass.

PR #352, opened 2026-08-24
feat(admin): send targetUser when creating a service account

### LINKED ISSUE #340: admin service-account create cannot target another user (targetUser unsupported in CLI and rc-core)
## Summary

The RustFS server's `add-service-account` admin API accepts a `targetUser` field, which lets an owner/root caller create a service account parented to **another** IAM user. The Console uses this. `rc` cannot: neither the CLI nor `rc-core` ever sends `targetUser`, so `rc admin service-account create` can only ever mint a key parented to whatever identity the alias authenticates as.

The result is that a key created in the Console under user `test-user` and a key created with `rc` using an admin alias are parented differently, with no way to express the Console's behaviour from the CLI.

Same shape as #261 ("Allow to update the service account policy via CLI since RustFS Console allows").

## Reproduction

With alias `openprojectx` holding the root credentials (root access key `rustfs`) and an IAM user `test-user`, two keys created for `test-user` in the Console and two created with `rc`:

```console
$ rc admin service-account list openprojectx --user test-user
  test-user-ak (parent: test-user) [on]     # created in the Console
  test-user-ak2 (parent: test-user) [on]    # created in the Console

$ rc admin service-account list openprojectx
  test-ak (parent: rustfs) [on]             # created with rc
  test-ak2 (parent: rustfs) [on]            # created with rc
  test-ak-from-ui (parent: rustfs) [on]
```

Listing works fine once `--user` is passed. The gap is creation: `rc admin service-account create` takes an alias, an access key and a secret key, and exposes `--name`, `--description`, `--policy`, `--policy-json` and `--expiry` — but nothing that selects a parent user, so every key it creates is parented to the alias identity (`rustfs` above). There is no invocation that produces a `parent: test-user` key.

For contrast, the Console issues `GET /rustfs/admin/v3/list-service-accounts?user=test-user` and a create carrying `"targetUser": "test-user"`, and both work against this same server.

## Why this looks like a gap rather than intended behaviour

**The server supports it.** `add_service_account_parent_within_scope` in `rustfs/src/admin/handlers/service_account.rs` gates the field as:

```rust
owner || target_user == req_user || target_user == req_parent_user
```

so an owner (root credential) may target any existing user, while a non-owner stays confined to its own scope (GHSA-5354). The handler reads `create_req.target_user` and falls back to `cred.access_key` only when it is absent.

**The docs document it.** `security-compliance/iam/sts.md` lists `targetUser` in the creation body and notes: *"`targetUser` requires admin privileges; regular users create service accounts for themselves."*

**The client omits it.** `targetUser` / `target_user` does not appear anywhere in this repository:

- `crates/core/src/admin/types.rs:405` — `CreateServiceAccountRequest` has `policy`, `expiry`, `name`, `description`, `access_key`, `secret_key`, and no target-user field, so the serialized body can never contain one.
- `crates/cli/src/commands/admin/service_account.rs` — `CreateArgs` has no user flag. `ListArgs` does have `--user` (line 46), so listing another user's accounts already works; only creation is missing.

Verified against `main` @ ef82a1d and published `rc-core` / `rc-s3` 0.1.31, `rc 0.1.31`.

## Suggested fix

1. Add `pub target_user: Option<String>` to `CreateServiceAccountRequest`, serialized as `#[serde(rename = "targetUser", skip_serializing_if = "Option::is_none")]` so existing callers are unaffected.
2. Add a `--user <USER>` flag to `CreateArgs` and pass it through, mirroring the existing `--user` on `service-account list`.

Happy to open a PR if the approach looks right.

## Motivation

We build a Kubernetes operator on top of `rc-core`/`rc-s3` that reconciles RustFS buckets, users, policies and access keys. Because `targetUser` is unreachable from the client, issuing a key for a managed user forces the operator to re-authenticate **as that user**, which means it must hold the user's password, and every managed user's policy has to grant `admin:CreateServiceAccount`, `admin:ListServiceAccounts` and `admin:RemoveServiceAccount` just so it can manage its own keys. With `targetUser` the admin credential alone would be sufficient and none of that would be needed.

loverustfs:
## Related issue

Fixes https://github.com/rustfs/cli/issues/340

## Background

RustFS `add-service-account` accepts `targetUser` so an owner credential can create a service account parented to another IAM user. The Console already sends that field. `rc admin service-account create` did not, so every key was parented to the alias identity. Listing another user's keys with `--user` already worked.

Operators that reconcile keys for managed users currently have to re-authenticate as each user. With `targetUser`, the admin alias is enough.

## Solution

- Add `target_user: Option<String>` to `CreateServiceAccountRequest`, serialized as `targetUser` and omitted when unset.
- Add `--user <USER>` to `rc admin service-account create`, matching `ls --user`.
- Empty `--user` values are omitted so accidental blank flags do not send `targetUser: ""`.

This is additive. Existing create commands without `--user` serialize the same body as before.

This PR must be marked BREAKING because `docs/reference/rc/admin.md` is a protected CLI behavior contract. The change is additive; no JSON schema or config `schema_version` bump applies.

## Tests

- Core: `targetUser` is present only when set, and deserializes when present
- CLI: `--user` is forwarded; empty values are omitted
- Integration: mock admin server receives `targetUser`, omits it without `--user`, and omits it for `--user ""`
- Help contract includes create `--user`
- Regression: `scripts/regression/service-account-target-user.sh`

## Validation

```bash
cargo fmt --all
cargo test --workspace
./scripts/regression/service-account-target-user.sh
```

### DIFF
--- a/crates/cli/src/commands/admin/service_account.rs
+++ b/crates/cli/src/commands/admin/service_account.rs
@@ -76,6 +76,11 @@ pub struct CreateArgs {
     /// Optional expiration time (ISO 8601 format)
     #[arg(long)]
     pub expiry: Option<String>,
+
+    /// Parent IAM user. Owner credentials may create a service account for
+    /// another user; omitted requests stay parented to the alias identity.
+    #[arg(long)]
+    pub user: Option<String>,
 }
 
 #[derive(clap::Args, Debug)]
@@ -300,6 +305,11 @@ fn build_create_service_account_request(
         description: args.description.clone(),
         access_key: args.access_key.clone(),
         secret_key: args.secret_key.clone(),
+        target_user: args
+            .user
+            .as_ref()
+            .filter(|value| !value.is_empty())
+            .cloned(),
     }
 }
 
@@ -512,11 +522,13 @@ mod tests {
             policy: None,
             policy_json: None,
             expiry: None,
+            user: None,
         };
 
         let request = build_create_service_account_request(&args, None);
         assert_eq!(request.name.as_deref(), Some("AKIAIOSFODNN7EXAMPLE"));
         assert!(request.expiry.is_none());
+        assert!(request.target_user.is_none());
     }
 
     #[test]
@@ -530,6 +542,7 @@ mod tests {
             policy: None,
             policy_json: None,
             expiry: Some("2030-01-01T00:00:00Z".to_string()),
+            user: None,
         };
 
         let request = build_create_service_account_request(&args, Some("{\"x\":1}".to_string()));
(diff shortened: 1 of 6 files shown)


cxymds:
LGTM

PR #353, opened 2026-08-24
feat(cp): copy objects across aliases via download and upload

loverustfs:
## Background

`rc cp` already performs same-alias S3-to-S3 copies with server-side CopyObject, including recursive prefix copies. Copies between different aliases failed with `unsupported_feature`, even though `rc mirror` already downloads and re-uploads across aliases. Operators had to stage through a local path or switch to mirror.

## Solution

- Same alias: keep CopyObject / multipart copy.
- Different aliases: HEAD the source, download to a temporary file, upload with the destination alias credentials, then delete the staging file.
- Source content type and user metadata are copied unless `--metadata-directive replace` is set.
- Storage-class rejection for cross-alias copies uses the upload multipart threshold, because the write is an upload rather than CopyObject.
- Client cache now includes both source and destination aliases.

This is additive. Same-alias CopyObject behavior is unchanged.

This PR must be marked BREAKING because `docs/reference/rc/cp.md` is a protected CLI behavior contract. The change is additive; no JSON schema or config `schema_version` bump applies.

## Tests

- Unit: planner client aliases for cross-alias and same-alias copies, storage-class threshold, content-type and user-metadata copy vs replace
- Integration: recursive dry-run across aliases without CopyObject; live recursive copy uses LIST + GET + PUT; single-object overwrite uses GET then PUT and never sends `x-amz-copy-source`; upload forwards `x-amz-meta-*` from the source HEAD
- Regression: `scripts/regression/cross-alias-copy.sh`

## Validation

```bash
cargo fmt --all
cargo test --workspace
./scripts/regression/cross-alias-copy.sh
```

### DIFF
--- a/crates/cli/src/commands/cp.rs
+++ b/crates/cli/src/commands/cp.rs
@@ -7,10 +7,10 @@ use jiff::Timestamp;
 use rc_core::alias::RetryConfig;
 use rc_core::{
     AliasManager, Error, MetadataDirective, MultipartCopyCancellation, MultipartCopyOptions,
-    ObjectEncryptionRequest, ObjectInfo, ObjectStore as _, ObjectWriteOptions, ParsedPath,
-    RemotePath, SseCustomerKey, TransferCancellation, TransferCandidate, TransferControls,
-    TransferCopyOptions, TransferExecutor, TransferOutcomeState, TransferPlan, TransferReadOptions,
-    TransferSelection, parse_path,
+    ObjectAttributes, ObjectEncryptionRequest, ObjectInfo, ObjectStore as _, ObjectWriteOptions,
+    ParsedPath, RemotePath, SseCustomerKey, TransferCancellation, TransferCandidate,
+    TransferControls, TransferCopyOptions, TransferExecutor, TransferOutcomeState, TransferPlan,
+    TransferReadOptions, TransferSelection, parse_path,
 };
 use rc_s3::S3Client;
 use serde::Serialize;
@@ -1091,6 +1091,9 @@ fn validate_storage_class_plan(
         })?;
         let multipart = match &item.payload {
             CpOperation::LocalToRemote { .. } => size > MULTIPART_THRESHOLD,
+            CpOperation::RemoteToRemote { source, target, .. } if source.alias != target.alias => {
+                size > MULTIPART_THRESHOLD
+            }
             CpOperation::RemoteToRemote { .. } => rc_core::requires_multipart_copy(size),
             CpOperation::RemoteToLocal { .. } => false,
         };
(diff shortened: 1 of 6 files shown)


crates/cli/src/commands/cp.rs
```diff
@@ -1309,6 +1324,127 @@ async fn perform_planned_remote_copy(
...
+                },
+                |copied, _total| on_progress(copied),
+            )
+            .await?;
+        if downloaded != planned_size {
+            return Err(Error::Conflict(format!(
+                "Source changed after copy planning: {source}"
+            )));
+        }
+        let options = piped_copy_write_options(args, &current, encryption)?;
```
cxymds, line 1405: [P2] This replace path is currently unreachable for the advertised cross-alias flow. validate_fidelity_directions rejects MetadataDirective::Replace for every Remote-to-Remote request before dispatch, so the cross-alias PUT can never execute these semantics even though the PR documentation and tests promise them. Please scope that rejection to the server-side CopyObject path, or allow REPLACE for cross-alias uploads and cover it with an end-to-end CLI test.

crates/cli/src/commands/cp.rs
```diff
@@ -1309,6 +1324,127 @@ async fn perform_planned_remote_copy(
...
+                &staging,
+                &TransferReadOptions {
+                    version_id: current.version_id.clone(),
+                    customer_key: args.source_customer_key.clone(),
+                    ..TransferReadOptions::default()
+                },
+                |copied, _total| on_progress(copied),
+            )
+            .await?;
+        if downloaded != planned_size {
```
cxymds, line 1385: [P2] The source is HEADed before the GET, but after the download this only checks the byte count. For an unversioned source, a concurrent same-size PUT can therefore be copied to the destination without being detected. Version IDs protect versioned objects, but the common unversioned path needs a post-download ETag/HEAD validation (or an If-Match-style read condition) before uploading.

crates/cli/src/commands/cp.rs
cxymds: Fixed in 8ec8c6e. MetadataDirective::Replace is now rejected only for same-alias server-side CopyObject; cross-alias transfers use the upload path. Added an end-to-end cross-alias replace test verifying replacement metadata and that no CopyObject request is used. All workspace checks pass.

crates/cli/src/commands/cp.rs
cxymds: Fixed in 8ec8c6e. After the cross-alias download, the source is revalidated with a version-aware and ETag-aware HEAD before the destination upload. Unversioned same-size replacements are now detected, with an end-to-end regression test. All workspace checks pass.

PR #354, opened 2026-08-25
feat(transfer): align cross-alias copy, move, and diff on source identity

loverustfs:
## Related work

Builds on #351 (mirror incremental identity) and #353 (cross-alias `cp`). Both are merged into this branch, so **review this after those two land**; the diff against `main` includes them.

Also resolves the `mv` half of the cross-alias gap and the `diff` comparison inconsistency.

## Background

#351 and #353 are each correct on their own, but together they leave a seam. A remote-to-remote transfer that streams through the client cannot preserve the source ETag — the destination computes its own, and multipart completion makes it differ even when the bytes are identical. #351 solved this for `mirror` by recording the source ETag in `x-amz-meta-rc-source-etag`. #353 added cross-alias `cp` without writing that key.

The result breaks the most common migration shape:

```bash
rc cp --recursive stage/data/ prod/archive/    # first full pass
rc mirror stage/data/ prod/archive/ --compare auto   # incremental
```

The second command re-copies the entire tree, because it cannot tell a faithful cross-alias copy from a changed object.

Two related asymmetries came out of the same audit:

- `rc mv` still rejected cross-alias remote moves with `unsupported_feature` even though `rc cp` now supports them, so the two commands disagreed about what is possible. Its S3-to-S3 path also duplicated copy logic instead of reusing `cp`, unlike its local↔remote paths.
- `rc diff` compared only size and the listed ETag, so it reported `≠` for exactly the pairs `mirror --compare auto` calls synchronized. Using `diff` to verify a `mirror` run gave contradictory answers.

## Root cause

Three commands each decided independently whether two remote objects hold the same data, and only one of them knew about the recorded source identity.

## Solution

A shared `object_identity` module owns the metadata key, the reader, and the writer, so the three commands cannot drift apart again. The first commit is a pure refactor with no behavior change; the second changes behavior.

- **`cp`**: cross-alias uploads record the source ETag. This is `rc` bookkeeping rather than user data, so it is written even with `--metadata-directive replace`, while that flag still drops source user metadata. Same-alias copies use server-side CopyObject, which preserves the ETag, and are untouched.
- **`mv`**: cross-alias remote moves now work by reusing the `cp` streaming path rather than reimplementing it, and inherit its constraints (SSE-C rejected, `--storage-class` rejected above the multipart threshold). The source is deleted only after its own copy succeeds, so a failed or partial move never destroys data that did not arrive. Same-alias moves keep server-side CopyObject.
- **`diff`**: adds `--compare auto|etag|size`, matching `rc mirror --compare`. `auto` reads the recorded identity through HeadObject, and only for same-size pairs whose listed ETags differ, so an unchanged tree costs no extra requests and `etag`/`size` never issue HeadObject.

## Behavior changes

- `rc diff` default moves from the old size-plus-ETag rule to `auto`. A target recording the source ETag is now `Same` instead of `Different`, so the command exits `0` where it previously exited `1`. Pass `--compare etag` to keep the previous comparison.
- `rc diff` no longer treats entries with unknown size on either side as equal; they are reported as different.
- `rc mv` between different aliases succeeds instead of returning the unsupported-feature exit code. Automation relying on that rejection to detect an unsupported operation must be updated.

This PR must be marked BREAKING because `docs/reference/rc/cp.md`, `mirror.md`, `mv.md`, and `diff.md` are protected CLI behavior contracts. No JSON schema or config `schema_version` bump applies.

## Tests

- Shared helper: read/write round trip, `x-amz-meta-` prefix and case variants, empty values rejected, unrelated metadata preserved
- `cp`: identity recorded, recorded under `--metadata-directive replace`, omitted when the source has no ETag
- `diff`: each compare mode, mismatched identity, size mismatch and unknown size in all three modes, HeadObject lookup gating, `--diff-only`
- Integration: cross-alias `cp` upload carries the metadata header; cross-alias `mv` does LIST/GET/PUT/DELETE and never sends `x-amz-copy-source`; a failed move takes the cross-alias path and leaves the source intact; `diff --compare auto` reports same via HeadObject while `etag` reports different with no HeadObject
- End-to-end: `cp --recursive` across aliases, then `mirror --compare auto` skips instead of recopying — the exact gap this PR closes, with a mock that persists what the upload wrote
- Regression: `scripts/regression/cross-alias-identity.sh`

Every new assertion was verified against a deliberately mutated implementation (identity write disabled, cross-alias dispatch disabled, identity comparison disabled) to confirm it fails when the behavior regresses rather than passing unconditionally.

## Validation

```bash
cargo fmt --all --check
cargo clippy --workspace --all-targets -- -D warnings
cargo test --workspace
./scripts/regression/cross-alias-identity.sh
./scripts/regression/mirror-identity.sh
./scripts/regression/cross-alias-copy.sh
```

All 59 test binaries pass with zero failures, and the two prerequisite regression suites still pass unchanged.


crates/cli/src/commands/cp.rs
```diff
@@ -1309,6 +1325,163 @@ async fn perform_planned_remote_copy(
...
+        target_client,
+        source,
+        target,
+        &source_info,
+        encryption,
+        &ignore_progress,
+        &args,
+    )
+    .await?;
+    Ok(result.object)
```
cxymds: [P1] The cross-alias move path loses the source identity returned by the copy. copy_object_across_aliases returns only result.object, so mv cannot delete the exact version that was copied or conditionally delete the unversioned source. A source update after the copy HEAD/GET can leave the destination with an older object while the subsequent plain delete removes or hides the newer current source. Please return the copied source version/ETag and use version-aware or conditional deletion in mv.

crates/cli/src/commands/cp.rs
```diff
@@ -1309,6 +1325,163 @@ async fn perform_planned_remote_copy(
...
+        {
+            attributes.user_metadata.clone_from(metadata);
+        }
+    }
+    // The destination computes its own ETag, so record the source ETag the same
+    // way `rc mirror` does. Without this a later `mirror --compare auto` cannot
+    // tell a faithful cross-alias copy from a changed object and recopies it.
+    // This is `rc` bookkeeping rather than user data, so it survives --metadata-directive replace.
+    if let Some(source_etag) = source.etag.as_deref() {
+        set_source_identity(&mut attributes, source_etag);
```
cxymds, line 1534: [P2] This identity-on-REPLACE branch is still unreachable from the CLI: the shared validate_fidelity_directions check above rejects metadata REPLACE for all Remote-to-Remote requests before this helper is called. The PR contract explicitly says the rc-source-etag bookkeeping survives --metadata-directive replace, so the validator needs to distinguish cross-alias PUTs from unsupported same-alias CopyObject semantics.

crates/cli/src/commands/cp.rs
```diff
@@ -1256,7 +1272,7 @@ async fn perform_planned_remote_copy(
                 "RustFS beta.10 does not persist storage class for multipart copies".to_string(),
             ));
         }
-        let current = client.head_object(source).await?;
+        let current = source_client.head_object(source).await?;
```
cxymds, line 1299: [P2] For objects above the 5 GiB single-copy limit, same-alias copies use multipart copy here, but no rc-source-etag identity is written. Standard S3 multipart completion commonly produces a composite ETag, so a later mirror or diff auto comparison can recopy an otherwise faithful copy. Either record the source identity for this path or add a backend-specific guarantee and regression test.

crates/cli/src/commands/diff.rs
```diff
@@ -302,24 +337,103 @@ async fn list_objects_map(
...
+        })
+        .map(|(key, _)| key.clone())
+        .collect();
+
+    for map_key in pending {
+        let Some(second_info) = second.get(&map_key) else {
+            continue;
+        };
+        let object_path = RemotePath::new(&path.alias, &path.bucket, &second_info.key);
+        if let Ok(info) = client.head_object(&object_path).await
```
cxymds: [P2] HeadObject is used only to fetch identity metadata; the listed size and ETag in second_info are left unchanged. If the object changes between LIST and this HEAD, auto comparison can combine the new identity with stale listed content metadata and report Same incorrectly. Please verify the HEAD size and ETag against the listed snapshot or treat the mismatch as a conflict.

cxymds:
Review note: this branch includes the identity race and cross-alias source-consistency issues identified in PRs #351 and #353. The P1 move deletion issue below makes the source race destructive.

crates/cli/src/commands/cp.rs
cxymds: Fixed in a2133b5. Cross-alias copy now returns the copied source version/ETag identity; mv deletes an exact version when available or uses If-Match for an unversioned source, and refuses deletion without identity. Same-alias moves pin the planned source version, and conditional-delete failure is covered by an integration test. All workspace checks pass.

crates/cli/src/commands/cp.rs
cxymds: Fixed in a2133b5. The fidelity validator now distinguishes same-alias CopyObject from cross-alias upload, so metadata REPLACE reaches the upload path and rc-source-etag bookkeeping survives it. The end-to-end cross-alias replace test covers this behavior. All workspace checks pass.

crates/cli/src/commands/cp.rs
cxymds: Fixed in a2133b5. Same-alias multipart-copy options now carry rc-source-etag, and multipart completion preserves that bookkeeping metadata even with MetadataDirective::Copy. Added an S3 request-level regression test for the create-multipart request. All workspace checks pass.

crates/cli/src/commands/diff.rs
cxymds: Fixed in a2133b5. Diff now compares HeadObject size and ETag with the listed snapshot and marks a mismatch or failed identity lookup as a conflict, so stale LIST metadata cannot produce Same. Added regression coverage for both size and ETag mismatches. All workspace checks pass.

cxymds:
Addressed in a2133b5. In addition to the inline fixes, this branch now includes the inherited mirror source preflight and cross-alias post-download revalidation. Move deletion is version/ETag conditional, and diff detects LIST/HEAD snapshot conflicts. Workspace checks pass.

PR #355, opened 2026-08-25
fix(ilm): preserve lifecycle extension fields

cxymds:
## Related Issues

Addresses rustfs/rustfs#6334, where `rc bucket lifecycle rule import` accepts MinIO-compatible lifecycle extensions but a subsequent export drops them.

## Summary of Changes

- Extend the core lifecycle model and XML codec to preserve `ExpiredObjectAllVersions`, `DelMarkerExpiration`, object-size predicates, repeated current/noncurrent transitions, and `NewerNoncurrentVersions`.
- Reject contradictory compatibility representations, including explicit `DelMarkerExpiration: false` combined with an enabled form.
- Validate lifecycle XML roots and successful response bodies so 2xx S3 `<Error>` envelopes cannot be treated as empty configurations.
- Add bounded alias-configured retries for lifecycle GET/PUT, 301/307/308 endpoint redirect handling, region refresh from `x-amz-bucket-region`, and fresh SigV4 signing for every request.
- Keep existing CLI read-modify-write operations lossless for extended lifecycle rules.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test --workspace`
- `make pre-commit` (workspace formatting, Clippy, tests, and doctests)
- `CARGO_BUILD_JOBS=2 cargo test -p rc-core --lib lifecycle::tests --no-fail-fast`
- `CARGO_BUILD_JOBS=2 cargo test -p rc-s3 --lib lifecycle --no-fail-fast`
- `CARGO_BUILD_JOBS=2 cargo test -p rustfs-cli --lib lifecycle --no-fail-fast`
- Manual import/export verification against the previously built RustFS main binary at commit `017ffb92f`; the fixed client preserves both lifecycle extensions.

### Adversarial review verdicts

- Correctness: PASS — success envelopes, filter predicates, action multiplicity, and compatibility conflicts are validated or preserved.
- Test coverage: PASS — focused regressions cover each review finding plus retry, dropped connections, redirects, and XML round-trips; workspace tests pass.
- Compatibility: PASS — legacy prefix/JSON shapes remain accepted; additive fields preserve newer S3/RustFS configurations.
- Security: PASS — XML roots are validated; redirects reject userinfo, HTTPS downgrade, untrusted hosts, and port changes; cross-authority custom x-amz headers are not forwarded.
- Concurrency / durability: PASS — PUT retries replay the exact idempotent body and re-sign each attempt; no partial local state is committed.
- Performance: PASS — lifecycle payloads are bounded configuration XML; retry/redirect loops are bounded and XML parsing overhead is linear.
- Simplicity: PASS — transport concerns are isolated in `lifecycle_xml_request`/`xml_request_once`; no unrelated paths changed.

## Impact

- Lifecycle import, export, list, and update now round-trip RustFS/MinIO delete-all and delete-marker expiration semantics, object-size filters, and all standard S3 lifecycle actions without loss.
- Existing standard S3 filters/actions remain supported.
- No RustFS server code, S3 object data, or output schema files are changed.

## Risks / Rollback

- Lifecycle GET/PUT uses a small raw XML path because the AWS SDK model does not expose the RustFS/MinIO extensions; requests remain SigV4-signed, Content-MD5 protected for PUT, and path/DNS-style aware.
- Redirect targets are restricted to the configured host or recognized S3 provider domains; custom x-amz headers are intentionally withheld across provider authority changes to avoid leaking credential-bearing values.
- Rollback is a normal client rollback by reverting the new commit; no server-side migration is required.

### DIFF
--- a/crates/core/src/lifecycle.rs
+++ b/crates/core/src/lifecycle.rs
@@ -34,6 +34,22 @@ pub struct LifecycleRule {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub tags: Option<HashMap<String, String>>,
 
+    /// Minimum object size in bytes for the rule filter.
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        alias = "ObjectSizeGreaterThan",
+        alias = "object_size_greater_than"
+    )]
+    pub object_size_greater_than: Option<i64>,
+
+    /// Maximum object size in bytes for the rule filter.
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        alias = "ObjectSizeLessThan",
+        alias = "object_size_less_than"
+    )]
+    pub object_size_less_than: Option<i64>,
+
     /// Expiration settings for current object versions
     #[serde(skip_serializing_if = "Option::is_none")]
     pub expiration: Option<LifecycleExpiration>,
@@ -46,6 +62,13 @@ pub struct LifecycleRule {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub transition: Option<LifecycleTransition>,
 
+    /// Additional transition settings for current object versions.
+    ///
+    /// `transition` remains the compatibility field for the first action;
+    /// this collection carries any subsequent actions returned by S3.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub transitions: Vec<LifecycleTransition>,
+
     /// Expiration settings for noncurrent object versions
     #[serde(skip_serializing_if = "Option::is_none")]
     pub noncurrent_version_expiration: Option<NoncurrentVersionExpiration>,
@@ -53,6 +76,10 @@ pub struct LifecycleRule {
     /// Transition settings for noncurrent object versions
     #[serde(skip_serializing_if = "Option::is_none")]
     pub noncurrent_version_transition: Option<NoncurrentVersionTransition>,
+
+    /// Additional transition settings for noncurrent object versions.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub noncurrent_version_transitions: Vec<NoncurrentVersionTransition>,
 
     /// Days after initiation to abort incomplete multipart uploads
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -74,14 +101,26 @@ struct LifecycleRuleInput {
     prefix: Option<String>,
     #[serde(default, alias = "Tags")]
     tags: Option<HashMap<String, String>>,
+    #[serde(
+        default,
+        alias = "ObjectSizeGreaterThan",
+        alias = "object_size_greater_than"
+    )]
+    object_size_greater_than: Option<i64>,
+    #[serde(default, alias = "ObjectSizeLessThan", alias = "object_size_less_than")]
+    object_size_less_than: Option<i64>,
     #[serde(default, alias = "Expiration")]
     expiration: Option<LifecycleExpirationInput>,
     #[serde(default, alias = "Transition")]
     transition: Option<LifecycleTransition>,
+    #[serde(default, alias = "Transitions")]
+    transitions: Vec<LifecycleTransition>,
     #[serde(default, alias = "NoncurrentVersionExpiration")]
     noncurrent_version_expiration: Option<NoncurrentVersionExpiration>,
     #[serde(default, alias = "NoncurrentVersionTransition")]
     noncurrent_version_transition: Option<NoncurrentVersionTransition>,
+    #[serde(default, alias = "NoncurrentVersionTransitions")]
+    noncurrent_version_transitions: Vec<NoncurrentVersionTransition>,
     #[serde(default, alias = "AbortIncompleteMultipartUploadDays")]
     abort_incomplete_multipart_upload_days: Option<i32>,
     #[serde(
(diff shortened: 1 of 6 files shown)


crates/s3/src/lifecycle_xml.rs
```diff
@@ -0,0 +1,492 @@
...
+    days_after_initiation: Option<i32>,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+struct LifecycleDelMarkerExpirationXml {
+    days: Option<i32>,
+}
+
+pub(crate) fn parse_lifecycle_configuration_xml(body: &str) -> Result<Vec<LifecycleRule>> {
```
cxymds, line 111: [P1] Reject S3 error envelopes before parsing\n\n`xml_request_inner` treats any 2xx response as success, and this parser accepts any root element while defaulting a missing `Rule` list to empty. A `200 <Error><Code>...</Code></Error>` response therefore becomes `Ok([])` (reproduced locally); `list`/`export` show no rules, and read-modify-write commands can erase the existing configuration, while PUT ignores the body and reports success. The SDK path being replaced explicitly checks for an `Error` root for both lifecycle operations. Please detect the S3 `Error` envelope (and validate the expected root) before returning success, with GET/PUT regression tests.

crates/s3/src/lifecycle_xml.rs
```diff
@@ -0,0 +1,492 @@
...
+    noncurrent_version_expiration: Option<NoncurrentVersionExpirationXml>,
+    #[serde(rename = "NoncurrentVersionTransition", default)]
+    noncurrent_version_transitions: Vec<NoncurrentVersionTransitionXml>,
+    abort_incomplete_multipart_upload: Option<AbortIncompleteMultipartUploadXml>,
+    del_marker_expiration: Option<LifecycleDelMarkerExpirationXml>,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+struct LifecycleFilterXml {
```
cxymds, line 42: [P1] Preserve object-size predicates during lifecycle round-trips

`LifecycleRuleFilter` supports `ObjectSizeGreaterThan` and `ObjectSizeLessThan`, including inside `And`, but this DTO has no corresponding fields and serde silently ignores them. A rule such as Prefix + 500..64000 is parsed as Prefix-only (reproduced locally), so export or any add/edit/remove-by-id read-modify-write sends a broader rule and can expire or transition objects outside the original range. Please carry these predicates through the core model/XML codec, or fail explicitly when an unsupported filter is returned. See the S3 filter contract: https://docs.aws.amazon.com/AmazonS3/latest/API/API_LifecycleRuleFilter.html

crates/s3/src/lifecycle_xml.rs
```diff
@@ -0,0 +1,492 @@
...
+                expired_object_all_versions: expiration.expired_object_all_versions,
+            }),
+            expiration
+                .expired_object_delete_marker
+                .filter(|value| *value),
+        ),
+        None => (None, None),
+    };
+
+    let transition = rule
```
cxymds: [P1] Do not truncate repeated lifecycle actions

S3 permits one or more lifecycle transition and expiration actions, but this conversion keeps only the first `Transition` and `NoncurrentVersionTransition`. It also omits `NewerNoncurrentVersions` from noncurrent transitions. On any read-modify-write, later transitions and version-retention limits are silently deleted. Please model these as collections/full fields or reject configurations that cannot be represented instead of round-tripping them lossily. The API documents the repeated actions and fields here: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLifecycleConfiguration.html

crates/s3/src/client.rs
```diff
@@ -6294,232 +6316,32 @@ impl ObjectStore for S3Client {
...
     async fn get_bucket_lifecycle(&self, bucket: &str) -> Result<Vec<LifecycleRule>> {
-        let response = match self
-            .inner
-            .get_bucket_lifecycle_configuration()
-            .bucket(bucket)
-            .send()
-            .await
-        {
-            Ok(resp) => resp,
+        let url = self.lifecycle_url(bucket)?;
```
cxymds, line 6543: [P2] Preserve configured retries and endpoint-redirect handling

Lifecycle GET/PUT now use the separately constructed `xml_http_client`; `alias.retry` is applied only to the SDK `config_loader`, and this client is configured with `Policy::none()` for redirects. Consequently lifecycle requests make one attempt and fail on transient network/5xx errors or S3 301/307 endpoint redirects, even when the alias is configured to retry. Please reuse the SDK transport or add bounded retries plus endpoint/region re-resolution with a fresh signature per attempt. S3's REST API guidance requires clients to handle redirects: https://docs.aws.amazon.com/AmazonS3/latest/developerguide/RESTRedirect.html

crates/core/src/lifecycle.rs
```diff
@@ -58,6 +63,141 @@ pub struct LifecycleRule {
...
+    }
+}
+
+fn normalize_del_marker_input(
+    input: Option<LifecycleDelMarkerInput>,
+    fallback_days: Option<i32>,
+) -> std::result::Result<Option<LifecycleDelMarkerExpiration>, String> {
+    match input {
+        None => Ok(None),
+        Some(LifecycleDelMarkerInput::Flag(false)) => Ok(None),
```
cxymds: [P2] Preserve explicit false until compatibility conflict validation

`normalize_del_marker_input` maps `Flag(false)` to `None`, so an explicitly supplied nested `DelMarkerExpiration: false` is indistinguishable from omission. An input containing that false plus a top-level `delMarkerExpiration` object/true is therefore accepted and enables marker expiration (reproduced locally), despite the PR's documented “reject ambiguous values” behavior. Preserve presence/false separately and reject contradictory representations.

crates/s3/src/lifecycle_xml.rs
cxymds: Fixed in 9eb6b79: the core and XML models now retain direct and And ObjectSizeGreaterThan/ObjectSizeLessThan predicates, and the builder emits them in the corresponding filter form. Added round-trip coverage for size filters.

crates/s3/src/client.rs
cxymds: Fixed in 9eb6b79: lifecycle XML requests now use alias retry settings for network, 429, and 5xx failures; follow bounded 301/307/308 redirects; refresh the signing region from x-amz-bucket-region; and re-sign each redirected request. Added retry and redirect tests.

crates/s3/src/lifecycle_xml.rs
cxymds: Fixed in 9eb6b79: current and noncurrent transition actions are retained beyond the compatibility first-action fields, and noncurrent transitions carry NewerNoncurrentVersions. The XML builder re-emits every action, with round-trip coverage.

crates/core/src/lifecycle.rs
cxymds: Fixed in 9eb6b79: compatibility normalization preserves explicit false presence until both representations are compared, rejecting false plus enabled conflicts while keeping standalone false compatible. Added conflict regressions.

crates/s3/src/lifecycle_xml.rs
cxymds: Fixed in 9eb6b79: lifecycle GET now validates the LifecycleConfiguration root before conversion, and PUT validates successful response XML. 2xx Error envelopes and malformed bodies now fail instead of being treated as an empty or successful configuration. Added GET and PUT regression tests.

PR #359, opened 2026-08-26
docs(admin): document account exit codes; ci: narrow the protected-files marker

sineld:
## Summary

Follow-up to #358, two independent pieces that came out of reviewing it.

**1. The account and two-factor commands now say how their failures are classified.** Every other workflow section in `docs/reference/rc/admin.md` does this — "Permission failures return the authentication exit code", "conflicts, and network failures retain distinct error classes and exit codes" — and this section only mentioned the usage error for a prompt it could not offer.

**2. The protected-files check no longer demands the `BREAKING` marker for a purely additive reference change.** #358 hit this: it documented new commands, which is an addition, and `AGENTS.md` asks for exactly that as step 2 of the Breaking Change process — so the check required the PR to assert in its description that it was breaking, which it was not.

## The exit-code table

| Condition | Code |
|---|---|
| Wrong verification code, recovery code, or password | authentication (4) |
| Locked out after repeated wrong codes | network (3) |
| At-rest protection not configured on the server | unsupported (7) |
| Route absent, on a server predating these endpoints | not found (5) |
| Recovery-code output path already occupied | conflict (6) |
| Missing, conflicting, or unprompted flags | usage (2) |
| Already enabled, not enabled, or no pending enrollment | general (1) |

Each row was read off `AdminClient::map_error` and the server's `map_service_error` rather than assumed. Two of them are now pinned by tests, because they are the rows somebody will build retry logic on:

- **The lockout is retryable, but not immediately.** It arrives as the network code, which normally means "retry", and the underlying `SlowDown` means "back off". A script that retries at once is simply refused again. `a_lockout_is_reported_as_retryable_rather_than_a_refusal`.
- **An absent route is not-found, not unsupported.** The unsupported code is what an up-to-date server sends when at-rest protection is unconfigured. Two different situations, two different codes. `an_absent_route_is_reported_as_not_found`.

While pinning the second one I found `mfa_enroll_reports_unsupported_when_the_server_has_no_such_route` was named for a route that does not exist, but drives a 501 — a different failure. Renamed to say what it checks.

Also documented: a failed `--output-file` write is the one case where a non-zero exit does not mean nothing happened. The server has already issued the set, so the codes are printed and the exit code reports only that the file was not written.

## The check change

Reference paths need the marker only when existing lines move — a rewritten sentence, a removed flag, a deleted page. The schema, exit-code and config files keep the old rule: a new exit code or config field extends a contract even when nothing existing moves, which is precisely what the marker is for.

Two deliberate calls:

- **Rename detection is off**, so moving a reference page counts as a rewrite. Where a reader or an inbound link lands does change.
- **Only `docs/reference/rc/` is relaxed.** The relaxation is about prose that documents new surface, not about code.

The job also collects failures now instead of exiting on the first, so a PR touching two protected paths learns about both in one run.

### Verified locally before sending

Shell logic in CI is easy to get subtly wrong, so I ran the extracted script against real commits in this repo:

| Scenario | Expected | Result |
|---|---|---|
| Reference changed by addition only, no marker | pass | `exit=0` |
| Existing reference line reworded, no marker | fail | `exit=1` |
| Same, with the marker | pass | `exit=0` |
| Additive change to `exit_code.rs`, no marker | fail | `exit=1` |

The last one also confirms the collect-don't-bail change: it reported the `exit_code.rs` failure *and* that the reference passed, in one run.

## Testing

```
cargo fmt --all --check                                  # clean
cargo clippy --workspace --all-targets -- -D warnings    # clean
cargo test --workspace                                   # 2525 tests, 0 failed
```

## Contracts

- `docs/reference/rc/admin.md` — additive: a new "Exit codes" subsection, no existing line touched. `git diff --numstat` reports `25` added, `0` removed. Under the check as this PR leaves it, that needs no `BREAKING` marker, which is the point of the second commit.
- `schemas/output_v1.json`, `crates/cli/src/exit_code.rs`, `crates/core/src/config.rs` — unchanged.

One thing worth flagging: for `pull_request` events GitHub runs the workflow from the merge ref, so I expect the *revised* check to run on this PR and pass without a marker. If it runs the version on `main` instead and goes red, say so and I will add the marker to this description.

The two commits are independent — happy to split them if you would rather take one and not the other.


.github/workflows/ci.yml
```diff
@@ -101,28 +101,82 @@ jobs:
...
+          # the Breaking Change process, in fact. Demanding the marker for that
+          # makes every additive PR claim to be breaking, so these paths need it
+          # only when existing lines move: a rewritten sentence, a removed flag,
+          # a deleted or renamed page.
+          ADDITIVE_OK_FILES=(
+            "docs/reference/rc/"
+          )
+
+          BASE="origin/${{ github.base_ref }}"
+          CHANGED_FILES=$(git diff --name-only "$BASE"...HEAD)
```
overtrue: [P2] Disable rename detection when collecting paths too. With a pure rename inside docs/reference/rc/, this command returns only the new path. The later --no-renames diff then sees an addition with zero deletions, so the check passes without BREAKING. I reproduced this with old.md renamed to new.md: the PR script exits 0; adding --no-renames here makes it exit 1 as intended. Please add that flag and a rename regression case.

docs/reference/rc/admin.md
```diff
@@ -191,6 +191,31 @@ asks for confirmation; `--yes` is required in `--json` mode or when stdin is not
...
 
+### Exit codes
+
+Every other workflow in this reference states how its failures are classified, so
+these do too. The account and two-factor commands map the server's answer onto
+the standard codes:
+
+| Condition | Code |
+|---|---|
+| Wrong verification code, recovery code, or password | authentication (4) |
```
overtrue: [P2] Split password-change failures from MFA authentication failures. For account passwd, a wrong current password returns InvalidRequest (HTTP 400) from the server's account handler, which AdminClient::map_error maps to General, exit 1. MFA disable uses AccessDenied instead, so exit 4 applies there. The combined row will mislead scripts handling password-change errors. Please document the distinction and add a test using the actual password endpoint's error response.

overtrue (changes requested):
Two issues to fix: renamed reference pages can bypass the marker check, and the password-error row does not match the server response.

.github/workflows/ci.yml
sineld: Confirmed and fixed in d6e52a9. I reproduced your case before changing anything: with a pure rename inside `docs/reference/rc/`, `git diff --name-only` returns only `new.md`, the per-file `--numstat --no-renames` on that single path sees 0 deletions, and the check reports "additive only" and exits 0.

`--no-renames` now applies to the collection step as well. That also closes a second hole I found while writing the test: with rename detection on, moving a file *out* of a protected directory (say `schemas/output_v1.json` → `schemas/output_v1.json.old`) reported only the destination, so nothing matched the protected prefix and the marker was never demanded. With the flag, the old path is listed and the check fires.

For the regression case: the logic was inline in `ci.yml`, so there was nothing to call. I extracted it to `scripts/check-protected-files.sh` (the workflow step is now a one-liner) and added `scripts/regression/protected-files.sh` following the convention of the other scripts in that directory. It builds synthetic histories in a temp dir and covers rename (with and without the marker), pure addition, an edited line, a deletion, and the protected-file-escape case above.

Verified it is genuinely red without the fix — removing `--no-renames` again fails exactly the two rename assertions and passes the other four.

docs/reference/rc/admin.md
sineld: You are right, and I traced it to both ends before changing the row. Fixed in aba3663.

Server side, `rustfs/src/admin/handlers/account.rs` answers a wrong current secret with `S3ErrorCode::InvalidRequest`, and the comment there says why: it reuses the message the new-password validation failures use, so a caller cannot distinguish "wrong current password" from "new password rejected" by probing. `rustfs/src/admin/handlers/mfa.rs` has no such ambiguity to protect and uses `AccessDenied`.

Client side, `AdminClient::map_error` sends `FORBIDDEN | UNAUTHORIZED` to `Error::Auth` (4) and `BAD_REQUEST` to `Error::General` (1), and `/account/password` goes through `account_response()`, which is plain `map_error` with no account-specific override. So the two commands really do report 1 and 4 for the same user mistake.

The table now has three rows instead of one — verification/recovery codes (4), password on `mfa disable` (4), current password on `account passwd` (1) — with a paragraph explaining the split so a reader does not treat it as an inconsistency to be normalised away.

Two tests, both driving the real endpoints through the mock admin server: `a_wrong_current_password_is_a_general_error_not_an_authentication_one` posts to `/account/password` with the handler's actual `400 InvalidRequest` body and asserts exit 1, and `a_wrong_password_on_mfa_disable_is_an_authentication_error` pins the `403 AccessDenied` counterpart at exit 4 so the pair cannot silently converge later.

`cargo fmt --all`, `cargo clippy --workspace -- -D warnings` and `cargo test --workspace` are all clean; `admin_account` is 20/20.

PR #371, opened 2026-09-09
fix(sql): align Select scan ranges with RustFS

GatewayJ:
## Related issues

## Background and user impact

rc sql rejects Parquet scan ranges that current RustFS supports, while allowing non-noop scan ranges for GZIP and BZIP2 input that RustFS rejects. This makes client-side behavior inconsistent with the target backend.

## Root cause

The CLI and S3 adapter each maintain their own scan-range rules, and both encode an outdated format matrix.

## Solution

- Introduce one rc-core validation boundary for scan ranges and input serialization.
- Allow Parquet scan ranges.
- Reject non-noop scan ranges for compressed input before network access.
- Preserve RustFS's accepted start=0, end=none no-op form for compressed input.
- Preserve existing JSON Document, negative-bound, and reversed-bound validation.
- Add core, CLI, and S3 adapter regression tests.

## Validation

- [x] cargo fmt --all --check
- [x] cargo test --workspace
- [x] Live RustFS Parquet query with a non-empty full-object scan range
- [x] Compressed scan range rejected locally with exit code 2

## Gate compatibility note

### DIFF
--- a/crates/cli/src/commands/sql.rs
+++ b/crates/cli/src/commands/sql.rs
@@ -343,26 +343,16 @@ fn validate_record_delimiter(name: &str, value: Option<&str>) -> std::result::Re
 }
 
 fn validate_scan_range_args(args: &SqlArgs) -> std::result::Result<(), String> {
-    if args.scan_start.is_none() && args.scan_end.is_none() {
-        return Ok(());
+    SelectScanRangeOptions {
+        start: args.scan_start,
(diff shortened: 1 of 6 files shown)


GatewayJ:
Closing because #370 was merged while this split was being prepared. #370 already contains the ScanRange compatibility fix; merging this branch would duplicate and conflict with main.

PR #372, opened 2026-09-09
feat(sql): support CSV input record delimiters

GatewayJ:
## Related issues

## Background and user impact

RustFS supports one- and two-byte CSV input record delimiters, but rc sql cannot currently express the S3 Select CSVInput RecordDelimiter field. Objects using CRLF or custom two-byte record boundaries therefore cannot be queried correctly from the CLI.

## Root cause

The core CSV input model and CLI argument set omit the input record delimiter, so the S3 adapter cannot serialize it into SelectObjectContent requests.

## Solution

- Add the additive --csv-input-record-delimiter option.
- Carry the value through rc-core without introducing AWS SDK types across the boundary.
- Validate the RustFS-compatible one- or two-byte range in both CLI preflight and the S3 adapter.
- Serialize the value as CSVInput.RecordDelimiter.
- Add parser, exit-code, validation, and SDK mapping tests.

## Validation

- [x] cargo fmt --all --check
- [x] cargo test --workspace
- [x] Live RustFS query using a custom two-byte record delimiter

## Gate compatibility note

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -880,6 +880,28 @@ mod tests {
     }
 
     #[test]
+    fn cli_accepts_sql_csv_input_record_delimiter() {
+        let cli = Cli::try_parse_from([
+            "rc",
+            "sql",
+            "local/reports/data.csv",
+            "--query",
+            "SELECT * FROM S3Object",
+            "--csv-input-record-delimiter",
+            "^Y",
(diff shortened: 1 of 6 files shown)


GatewayJ:
Closing because #370 was merged while this split was being prepared. #370 already contains CSV input RecordDelimiter support; merging this branch would duplicate and conflict with main.

PR #373, opened 2026-09-09
fix(sql): preserve S3 Select service errors

GatewayJ:
## Related issues

## Background and user impact

Modeled RustFS S3 Select failures can collapse to the generic service error text, hiding actionable codes such as UnsupportedScanRangeInput, Busy, and SlowDown.

## Root cause

The Select error mapper prefers the SDK wrapper string instead of modeled service metadata and drops error codes for otherwise unknown service errors.

## Solution

- Prefer modeled service messages with the SDK string as fallback.
- Preserve service codes in user-facing diagnostics.
- Classify Busy and SlowDown as retryable network errors.
- Preserve UnsupportedScanRangeInput and other Select error details.
- Retain the existing NotImplemented fallback behavior.
- Add focused classifier regression tests, including generic service-error fallback text.

## Validation

- [x] cargo fmt --all --check
- [x] cargo test --workspace
- [x] Live RustFS parser failure retained ParseSelectFailure and its detailed message

## Gate compatibility note

### DIFF
--- a/crates/cli/src/commands/table/mod.rs
+++ b/crates/cli/src/commands/table/mod.rs
@@ -406,10 +406,10 @@ fn properties(values: Vec<String>) -> Result<BTreeMap<String, String>> {
     Ok(result)
 }
 fn require_string(body: &Value, field: &str) -> Result<()> {
-    if !body
+    if body
         .get(field)
         .and_then(Value::as_str)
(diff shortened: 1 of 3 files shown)


GatewayJ:
Closing because #370 was merged while this split was being prepared. #370 already contains the Select service-error reporting fix; merging this branch would duplicate and conflict with main.

